PR #180, opened 2026-04-14
Fix typo in AC0012 Diagnostic Message

rvanbekkum:
Fix typo in AC0012 diagnostic message:

`Access = Interna` -> `Access = Internal`


rvanbekkum:
PR Build is failing, but I don't think it's due to my proposed change.

Arthurvdv:
Jep, that's on me. It seems that the permissions for dorny/test-reporter needs some work for external PR's. Let me merge this, as I'm quite sure this change will work. Thanks for the PR by-the-way!

PR #217, opened 2026-04-24
fix: dynamic TargetFramework in test projects for net8.0/net10.0 CI matrix

Arthurvdv:
## Problem

The CI pipeline fails for all net8.0 BC DevTools versions (v16.x through v18.0.35.14686) with CS1705 assembly version mismatches. Net10.0 and netstandard2.0 versions are green.

**Root cause**: Test csprojs hardcoded `TargetFramework=net10.0` and `NavBinaryTfm=net10.0` for all non-legacy TFMs. When testing net8.0 BC DevTools versions:
- Analyzer DLLs from `bin/Release/net10.0/` (compiled against net10.0 BC DevTools) were referenced
- NuGet resolved the net10.0 RoslynTestKit DLL instead of net8.0
- Both referenced higher `Microsoft.Dynamics.Nav.CodeAnalysis` versions than the net8.0 BC DevTools being tested

## Fix

- **6 test csprojs**: Make `TargetFramework` and `NavBinaryTfm` derive from `NavTargetFramework` dynamically. Net8.0 matrix entries now compile as net8.0 and reference net8.0 binaries. Legacy TFMs still compile as net10.0 (unchanged).
- **build-test.yml**: Install BC DevTools to the matching TFM folder (`net8.0` for net8.0 tests) instead of hardcoded `net10.0`.
- **Instructions**: Updated project-overview to reflect the new behavior.

Local dev default remains `net10.0` (no change for developers).

Fixes: https://github.com/ALCops/Analyzers/actions/runs/24895077064


Arthurvdv:
Closing in favor of adding these changes to #216

Issue #380, opened 2026-07-03
[Bug]: LC0095 false positive when casing of interface and implementation method differs


MODUSCarstenScholling:
### Analyzer

LinterCop

### Issue type

False positive (diagnostic fires when it shouldn't)

### Rule ID

LC0095

### Description

LC0095 fires false positive when the casing of an interface method and implementation method differs.

### AL code to reproduce

```al
codeunit 50102 MyCodeunit implements MyInterface
{
    procedure Myprocedure(MyInteger: Integer; MyText: Text; MyDate: Date)
    begin
        MyText := 'Hello';
        MyDate := Today();
    end;
}

interface MyInterface
{
    procedure MyProcedure(MyInteger: Integer; MyText: Text; MyDate: Date);
}
```

### Additional context

Fix is ready and I will create a PR.

PR #381, opened 2026-07-03
#380 fix(lintercop): avoid false positives for lc0095 interface impl parameter casing

MODUSCarstenScholling:
#380 fix(lintercop): avoid false positives for lc0095 interface impl parameter casing


Arthurvdv:
Thanks for the PR! 

P.S. I've made a small change to use the `SemanticFacts.IsSameName` for name comparisons to align with a project convention.

Issue #392, opened 2026-07-09
Feature: Add PC0038 and LC0097 Return-Flow Diagnostics


MODUSCarstenScholling:
# Feature: Add PC0038 and LC0097 Return-Flow Diagnostics

## Summary

Introduce two new analyzer rules to improve return-flow correctness and consistency in AL code:

- **PC0038**: NotAllCodePathsReturnValue (info)
- **LC0097**: MixedExitAndNamedReturnAssignment (disabled by default)

## Problem

Current analysis does not fully cover:
- routines that can end without returning a value
- routines that mix `exit(...)` and named-return assignments

Both patterns reduce readability and increase risk of logic defects.

## Scope

- Add PC0038 in PlatformCop
- Add LC0097 in LinterCop
- Add IDs, descriptors, resources, and analyzer wiring
- Add/adjust tests with meaningful positive/negative coverage
- Documentation pages and index references are handled in a PR in the documentation repo

See discussion [Warning/error if procedure with return value doesn't contain exit() and doesn't assign to the named return value](https://github.com/ALCops/Analyzers/discussions/362#discussion-10311249).

PR #393, opened 2026-07-09
feat: add PC0038 and LC0097 return-flow diagnostics

MODUSCarstenScholling:
# Feat: add PC0038 and LC0097 return-flow diagnostics

This PR introduces two new analyzer features for return-flow quality:

- PC0038 NotAllCodePathsReturnValue
- LC0097 MixedExitAndNamedReturnAssignment (disabled by default)

## What is new

- Added a new PlatformCop rule PC0038 to detect methods where not all code paths return a value.
- Added a new LinterCop rule LC0097 to detect methods that mix exit style and named-return assignment style.

## Test coverage

- Added comprehensive rule-specific test suites for PC0038 and LC0097.

## Additional updates

- **Resource generation fix**: updated cop project files to ensure reliable analyzer `.resx` strongly-typed resource generation/design-time inclusion.
- **Task naming cleanup**: renamed VS Code tasks with a consistent `ALCops:` prefix.
- **Diagnostic display consistency**: adopted shared method display formatting in:
- Updated method display (signature) in **ProcedureRequiresDocumentation** analyzer diagnostics

## Outcome

This feature set strengthens return-flow correctness and consistency, reduces ambiguous control-flow patterns, and provides clearer guidance during development.

Implements #392 
See discussion [Warning/error if procedure with return value doesn't contain exit() and doesn't assign to the named return value](https://github.com/ALCops/Analyzers/discussions/362#discussion-10311249).

### DIFF
--- a/.github/instructions/instruction-maintenance.instructions.md
+++ b/.github/instructions/instruction-maintenance.instructions.md
@@ -107,3 +107,5 @@ Include: project structure, templates, step-by-step guides, API reference, commo
 | `pc0035-use-set-auto-calc-fields-for-loops` | rule-scoped | PC0035 rule |
 | `lc0095-parameter-not-referenced` | rule-scoped | LC0095 rule |
 | `ac0006-run-page-implement-page-management` | rule-scoped | AC0006 rule |
+| `pc0038-not-all-code-paths-return-value` | rule-scoped | PC0038 rule |
+| `lc0097-mixed-exit-and-named-return-assignment` | rule-scoped | LC0097 rule |
--- a/.github/instructions/lc0097-mixed-exit-and-named-return-assignment.instructions.md
+++ b/.github/instructions/lc0097-mixed-exit-and-named-return-assignment.instructions.md
@@ -46,9 +46,9 @@ Detects methods with a named return variable that mix both styles:
 ## Test coverage
 
 **HasDiagnostic (6 cases):** NamedAssignmentAndExitValue, NamedAssignmentAndExitWithoutValue, NamedAssignmentInIfAndExitInElse, NamedCaseAssignmentAndExit, NamedIfElseIfElseAssignmentAndExit, NamedNestedIfElseIfAssignmentAndExit.
-**NoDiagnostic (10 cases):** NamedOnlyAssignment, NamedOnlyExit, UnnamedExitAndLocalAssignment, TryFunctionExcluded, NamedCaseOnlyAssignments, NamedIfElseIfElseOnlyExit, NamedNestedIfElseIfOnlyAssignments, TriggerBooleanOnlyExit, TriggerBooleanCaseOnlyExit, TriggerBooleanNestedIfElseIfOnlyExit.
+**NoDiagnostic (11 cases):** NamedOnlyAssignment, NamedOnlyExit, UnnamedExitAndLocalAssignment, TryFunctionExcluded, NamedCaseOnlyAssignments, NamedIfElseIfElseOnlyExit, NamedNestedIfElseIfOnlyAssignments, TriggerBooleanOnlyExit, TriggerBooleanCaseOnlyExit, TriggerBooleanNestedIfElseIfOnlyExit, NamedFieldSameNameOnlyExit.
 
 ## Known issues
 
-- Assignment target detection prefers `ReturnValueReferenceExpression` but keeps symbol-name fallback for SDK representation differences.
+- Assignment target detection prefers `ReturnValueReferenceExpression` and falls back to symbol identity — the fallback requires the symbol's `Kind` to be `ReturnValue` so that field members whose names happen to match the return variable (e.g. `Buf.Result := 5;` when `Result` is also the return name) are not misidentified as return-variable assignments. See `IsNamedReturnTarget` in `ALCops.Common/Extensions/OperationExtensions.cs`.
 - Methods that only use `exit` (without return-variable assignments) are intentionally not flagged.
--- a/.github/instructions/lc0097-mixed-exit-and-named-return-assignment.instructions.md
+++ b/.github/instructions/lc0097-mixed-exit-and-named-return-assignment.instructions.md
@@ -1,0 +1,54 @@
+---
+applyTo: 'src/ALCops.LinterCop/**/MixedExitAndNamedReturnAssignment*'
+---
+
+# LC0097: MixedExitAndNamedReturnAssignment
+
+## Purpose
+
+Detects methods with a named return variable that mix both styles:
+- assignment to the named return variable, and
+- usage of `exit(...)` or `exit`.
+
+## Diagnostic properties
+
+| Property | Value |
+|----------|-------|
+| ID | LC0097 |
+| Category | Usage |
+| Severity | Warning |
+| Enabled | No (disabled by default) |
+| CodeFix | No |
+| Message | `{0} '{1}' mixes exit() with assignments to the named return variable. Use one return style consistently.` |
+
+## Design decisions
+
+| Decision | Rationale |
+|----------|-----------|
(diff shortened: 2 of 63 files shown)


Arthurvdv:
## Code review

Three verified findings (all reproduced against the actual analyzers, not just read from code):

### 1. PC0038 false positive: `Error()`-terminated branches are treated as fall-through (High)

`src/ALCops.PlatformCop/Analyzers/NotAllCodePathsReturnValue.cs:242` — the path analysis has no notion of flow-terminating calls. `Error(...)` / `ThrowError` fall into the `default:` case and pass states through unchanged, so the branch is treated as reaching the end of the method without a value:

```al
procedure Compute(Input: Integer): Integer
begin
    if Input = 1 then
        exit(10)
    else
        Error('unsupported');
end;
```

→ reports `PC0038: 'Compute(Integer)': not all code paths return a value`. The same happens for a `case` whose `else` branch calls `Error()`. This guard-clause shape is pervasive in real AL code, and PC0038 is enabled by default, so this will produce large-scale noise.

**Suggested fix:** treat invocation statements binding to the built-in `Error` / `ThrowError` methods as path terminators (return `ImmutableHashSet<bool>.Empty`).

### 2. PC0038 false positive: named return initialized via `var` parameter or member call not recognized (High)

`src/ALCops.PlatformCop/Analyzers/NotAllCodePathsReturnValue.cs:108` — only a direct `IAssignmentStatement` marks the named return value as assigned. Passing it as a `var` argument, or initializing it through a method call (e.g. `Customer.Get(No)` in `procedure GetCustomer(No: Code[20]) Customer: Record Customer`), still triggers the diagnostic:

```al
procedure Compute() Result: Integer
begin
    ComputeInto(Result);
end;

local procedure ComputeInto(var Value: Integer)
begin
    Value := 42;
end;
```

→ reports `PC0038: 'Compute()': not all code paths return a value`. The `Get()`-into-named-record-return and `var`-out patterns are among the most common AL idioms.

**Suggested fix:** also mark the state as assigned when the named return value is passed as a `var` argument or is the receiver of an invocation (conservatively), or default-disable the rule until these mutation channels are modeled.

### 3. `IsNamedReturnTarget` matches any symbol by name (Medium)

`src/ALCops.Common/Extensions/OperationExtensions.cs:55` — after the `ReturnValueReferenceExpression` kind check, the fallback compares only the *name* of whatever symbol the target resolves to. A record field sharing the return variable's name is then treated as an assignment to the named return value. Verified consequences:

- **PC0038 false negative:** `procedure Compute(var Buf: Record ProbeBuffer) Result: Integer begin Buf.Result := 5; end;` reports nothing, while the identical method assigning `Buf.Amount` correctly reports PC0038.
- **LC0097 false positive:** `begin Buf.Result := 5; exit(1); end;` reports LC0097 although the named return value is never assigned.

The fallback is provably unnecessary: a genuine assignment to the named return value binds as `ReturnValueReferenceExpression` with `SymbolKind.ReturnValue`, which the first check already handles.

**Suggested fix:** drop the name-based fallback, or restrict it to symbols of kind `ReturnValue`.

---

**Verification performed:** built the PR head for the CI target matrix including `netstandard2.1` (0 errors); ran the full PlatformCop (491), LinterCop (274), DocumentationCop (69) and Common (10) test suites (all pass); confirmed SDK interface shapes (`ICaseLineOperation.Statement`, `ICompoundAssignmentStatement : IAssignmentStatement` — compound assignments are handled correctly in both rules); checked EnumProvider usage, `IsObsolete()` calls, `GetSymbolSafe()` usage, ID/resx/descriptor/help-URI consistency, fixture markers, and the `.csproj`/`tasks.json` changes (benign design-time-build additions only).

**Verdict:** the infrastructure, conventions, and shared-extension refactoring are solid, and LC0097's core logic is sound apart from the shared name-matching helper. PC0038's control-flow analysis has two verified false-positive classes (`Error()`-terminated branches, non-assignment initialization of named returns) that will make a default-enabled warning noisy on idiomatic AL code — these should be fixed (or the rule default-disabled) before merging.

*Generated with Claude Fable 5*

Arthurvdv:
## Additional finding: `MethodDeclarationSyntaxExtensions.IsTryFunction()` can be replaced with the SDK's `IMethodSymbol.IsTryFunction`

The new syntax-based helper in `src/ALCops.Common/Extensions/MethodDeclarationSyntaxExtensions.cs` reimplements something the SDK already computes semantically.

**What the SDK exposes**

`IMethodSymbol.IsTryFunction` is a public property on the symbol interface. The compiler sets it during method binding via `HasAttribute(AttributeKind.TryFunction) && MethodKind == MethodKind.Method` (`SourceMethodOrTriggerSymbol`), and `ReferenceMethodSymbol` implements it for symbols loaded from `.app` references. Verified that `get_IsTryFunction` also exists in the **netstandard2.1** SDK DLL, so it is safe across all three CI targets — no reflection wrapper or `#if NETSTANDARD2_1` guard needed.

**Why it's better than the syntax helper**

1. **Semantic, not string-based.** It uses the compiler's bound `AttributeKind.TryFunction`, not name matching on syntax tokens. No quoted-identifier or casing edge cases.
2. **It's free at the call site.** Both analyzers already fetch `ctx.ContainingSymbol as IMethodSymbol` a few lines *after* the try-function check (e.g. `NotAllCodePathsReturnValue.cs:41`). Just reorder: get the method symbol first, then:
   ```csharp
   if (methodSymbol.IsTryFunction)
       return;
   ```
3. **Bonus semantic correctness:** the SDK also gives try-functions an implicit `Boolean` return value (`lazyReturnValue = ReturnValueSymbol.Boolean`), which is relevant context for PC0038's return-flow analysis.

**Suggested change:** drop `MethodDeclarationSyntaxExtensions.cs` entirely and use `methodSymbol.IsTryFunction` in both `NotAllCodePathsReturnValue` and `MixedExitAndNamedReturnAssignment`. Fewer helpers to maintain, and the SDK stays the single source of truth.

(For reference: `XmlDocumentationProcedureConsistency` in DocumentationCop does its own syntax-based TryFunction detection, but that one legitimately needs syntax because it reports a diagnostic at the attribute's location. That's not the case here.)

*Generated with Claude Fable 5*

Arthurvdv:
Hi @MODUSCarstenScholling,

Looking promosing so far. Did you have a chance to look at the additional finding about replacing `MethodDeclarationSyntaxExtensions.IsTryFunction()` with the SDK's `IMethodSymbol.IsTryFunction()` ?

I also went ahead and did a manual review of the rules against the Base App. They catch quite a few instances, great work!

|id|Amount|
|--|--|
|PC0039|1470|
|LC0097|849|

I have a last question on the PC0039 diagnostic; The default severity is set to `Warning`. I'm wondering if that's a bit too strict and whether `Info` would be a better default? 

```AL
local procedure GetFromBuffer(var TempBlob: Codeunit "Temp Blob"): Text
var
    InStream: InStream;
    Data: Text;
    Line: Text;
begin
    if not TempBlob.HasValue() then
        exit;
    TempBlob.CreateInStream(InStream);
    while InStream.ReadText(Line) > 0 do
        Data += Line;
    exit(Data);
end;
```

For example, in the code above, the diagnostic suggests replacing `exit` with `exit('')`. While that's a valid suggestion, I'm not sure it's severe enough to warrant a warning from the PlatformCop. Would it make sense to lower the default severity from `Warning` to `Info`?

MODUSCarstenScholling:
Info is OK and makes more sense. 
I can change this on Monday.

Regarding IsTryFunction: this is not working with netstandard. So I removed it again.

MODUSCarstenScholling:
I have lowered the severity to Info for both. This makes more sense when looking at your findings. LC0097 is still disabled.

Arthurvdv:
Thanks a lot for this @MODUSCarstenScholling 🤗

Appreciate how you handled the review feedback. Merging into `main` now (and will be part of the next release).

Issue #406, opened 2026-07-16
feat: add LC0098 - Event subscriber name does not match the configured template


MODUSCarstenScholling:
## feat: add LC0098 - Event subscriber name does not match the configured template

Adds LinterCop rule **LC0098 EventSubscriberNamingPattern** with CodeFix (**BatchFixer FixAll**).  
The rule reports when an event subscriber procedure name does not match the identifier derived from its [EventSubscriber] attribute using one configurable template.

Default template:

```text
{Event Source}_{Event Name}[_{Element Name}]
```

This default matches the AL Language extension **Find Event** output verbatim (raw source and element names, quoted when required). Teams already using the tooling default get zero-config alignment. Severity is **Info**, category **Naming**.

## Template syntax

- Token placeholders support source object name, event name, and element name across five styles: raw, PascalCase, camelCase, snake_case, kebab-case.
- Raw placeholders emit values verbatim with no splitting or casing transform.
- Non-raw placeholders use shared word splitting and rendering in ALCops.Common.
- Optional groups in brackets are emitted only when all inner token placeholders are non-empty.
- Unknown brace sequences are emitted verbatim; new placeholders can be added via the parser placeholder map without grammar changes.

## Acronym-aware rendering

For non-raw styles, rendering now produces an **accepted-name set**:

- Preferred name is always first and follows **original casing wins**.
- Additional accepted variants are optionally added when KnownAcronyms contains a differently cased entry for an uppercase-carrying source word with the same case-insensitive key.
- A diagnostic is reported only when the method name matches **none** of the accepted variants (ordinal comparison).
- The CodeFix always suggests the **preferred** name.

Per-word rules:

- ID (any casing) becomes Id.
- Two-letter all-uppercase words stay uppercase.
- Words containing uppercase preserve original casing shape (first letter normalized to upper as needed).
- All-lowercase registered words use registry canonical casing.
- All-lowercase unregistered words use first-upper fallback.

## Configuration

alcops.json properties:

- SubscriberNamingPattern: string, default null (built-in default template used)
- KnownAcronyms: merged with built-in defaults, user entries override on same case-insensitive key

KnownAcronyms semantics:

- For all-lowercase source words: defines preferred canonical casing.
- For uppercase-carrying source words: can add accepted variants alongside the original spelling.
- Preferred name remains original-casing form for uppercase-carrying sources; CodeFix suggestion remains that preferred form.

Example:

If source contains LCY and KnownAcronyms includes Lcy:

[...]

PR #408, opened 2026-07-17
feat: add LC0098 - Event subscriber name does not match the configured template

MODUSCarstenScholling:
## feat: add LC0098 - Event subscriber name does not match the configured template

Adds LinterCop rule **LC0098 EventSubscriberNamingPattern** with CodeFix (**BatchFixer FixAll**).  
The rule reports when an event subscriber procedure name does not match the identifier derived from its [EventSubscriber] attribute using one configurable template.

Default template:

```text
{Event Source}_{Event Name}[_{Element Name}]
```

This default matches the AL Language extension **Find Event** output verbatim (raw source and element names, quoted when required). Teams already using the tooling default get zero-config alignment. Severity is **Info**, category **Naming**.

## Template syntax

- Token placeholders support source object name, event name, and element name across five styles: raw, PascalCase, camelCase, snake_case, kebab-case.
- Raw placeholders emit values verbatim with no splitting or casing transform.
- Non-raw placeholders use shared word splitting and rendering in ALCops.Common.
- Optional groups in brackets are emitted only when all inner token placeholders are non-empty.
- Unknown brace sequences are emitted verbatim; new placeholders can be added via the parser placeholder map without grammar changes.

## Acronym-aware rendering

For non-raw styles, rendering now produces an **accepted-name set**:

- Preferred name is always first and follows **original casing wins**.
- Additional accepted variants are optionally added when KnownAcronyms contains a differently cased entry for an uppercase-carrying source word with the same case-insensitive key.
- A diagnostic is reported only when the method name matches **none** of the accepted variants (ordinal comparison).
- The CodeFix always suggests the **preferred** name.

Per-word rules:

- ID (any casing) becomes Id.
- Two-letter all-uppercase words stay uppercase.
- Words containing uppercase preserve original casing shape (first letter normalized to upper as needed).
- All-lowercase registered words use registry canonical casing.
- All-lowercase unregistered words use first-upper fallback.

## Configuration

alcops.json properties:

- SubscriberNamingPattern: string, default null (built-in default template used)
- KnownAcronyms: merged with built-in defaults, user entries override on same case-insensitive key

KnownAcronyms semantics:

- For all-lowercase source words: defines preferred canonical casing.
- For uppercase-carrying source words: can add accepted variants alongside the original spelling.
- Preferred name remains original-casing form for uppercase-carrying sources; CodeFix suggestion remains that preferred form.

Example:

If source contains LCY and KnownAcronyms includes Lcy:

- ...LCY is accepted and remains preferred
- ...Lcy is additionally accepted
- ...LcY is still rejected

## Safety guards

LC0098 suppresses diagnostics when rename would be unsafe or invalid:

- **AL304 length guard**: preferred name longer than 120 chars is skipped.
- **Duplicate-name collision guard**: skipped when a sibling already has preferred name or another subscriber would compute to the same preferred name.
- Unresolvable source objects and obsolete methods are skipped.

## Shared infrastructure in ALCops.Common

- AcronymRegistry: built-in list plus user merge/override behavior
- IdentifierCaseStyle: shared style enum
- IdentifierNameRenderer: splitting and casing renderer, including accepted-variant generation

## Interaction with LC0092

LC0098 and LC0092 are orthogonal: structural template validation vs regex naming validation.  
Both can report independently.

## CodeFix

Renames only subscriber declaration identifier via syntax token replacement.  
PreferredName is passed from analyzer to CodeFix through diagnostic properties, so CodeFix does not recompute naming input.  
Supports FixAll via BatchFixer.

## Compatibility

Full netstandard2.1 support; no net8-only SDK dependency required.

## Tests (updated)

- ALCops.Common.Test: renderer coverage extended to accepted-variant behavior (including cross-product and duplicate prevention).
- LC0098 tests extended with new fixtures for:
  - acceptance of original uppercase acronym spelling
  - acceptance of user-pinned variant alongside original spelling
  - rejection of third mixed variant

Implements #406


Arthurvdv (changes requested):
## PR #408 Review — LC0098 EventSubscriberNamingPattern

**Summary:** Solid implementation overall. Built on net10.0 and netstandard2.1, all 33 LC0098 analyzer/CodeFix tests pass, but **2 of the PR's own new unit tests fail**, and the collision guard is case-sensitive where the AL compiler is case-insensitive.

### 🔴 Critical — PR's own AcronymRegistry tests fail: duplicate case-insensitive keys in `DefaultAcronyms`
`src/ALCops.Common/Helpers/AcronymRegistry.cs:35,40`
`DefaultAcronyms` contains both `"Bom"`/`"BoM"` and `"Uom"`/`"UoM"`. The registry keys on `ToUpperInvariant()` with last-entry-wins, so `"Bom"`/`"Uom"` are dead entries. `dotnet test` on the PR head: **Failed: 2** (`Default_ExposesEveryEntryFromDefaultAcronyms`, `Create_WithNullUserList_ReturnsDefaultsOnly` — `Expected: "Bom" But was: "BoM"`).
**Fix:** pick one canonical casing per key, or adjust the tests, but as written the duplicates are pointless since only one survives `Create()`.

### 🟠 Major — Collision guard is case-sensitive; AL duplicate-method detection is case-insensitive
`src/ALCops.LinterCop/Analyzers/EventSubscriberNamingPattern.cs:176,187`
`WouldCollideInContainingType` compares sibling names with `StringComparison.Ordinal`. If a sibling is named `mypublisher_onsomething()` and the preferred name is `MyPublisher_OnSomething`, the guard misses it and applying the CodeFix produces a duplicate-method compile error. SDK's `MethodSymbol.HasSameSignature` uses `SemanticFacts.IsSameName` (case-insensitive), matching repo convention. (The ordinal check at line 74 is correct — that one intentionally enforces casing.)
**Fix:** use `SemanticFacts.IsSameName()` for both collision checks.

### 🔵 Minor — Resx description documents a wrong default template
`src/ALCops.LinterCop/ALCops.LinterCopAnalyzers.resx:430`
Says `'{Event Source}_{EventName}[{Element Name}]'`; actual default is `'{Event Source}_{Event Name}[_{Element Name}]'`. Placeholder spelling is semantically significant (PascalCase vs raw), and the `_` in the optional group matters. Schema description is correct.

### ✅ Conventions verified OK
Analyzer structure (matches LC0092/LC0097), EnumProvider usage, IsObsolete, attribute argument extraction (indices match SDK `EventSubscriber` order, quoted names covered), per-compilation caching/performance, template parser robustness (unclosed braces/brackets degrade gracefully, bounded cross-product), netstandard2.1 (builds clean, correct `#if` pattern), CodeFix structure (BatchFixer, properties handoff, trivia preserved), descriptor/resx/schema/settings patterns, all 33 LC0098 tests pass with valid markers.

---

MODUSCarstenScholling:
Addresses three review findings on LC0098 (EventSubscriberNamingPattern) and its shared `AcronymRegistry` infrastructure.

### 1. `AcronymRegistry`: multi-variant support per case-insensitive key

Previously, `DefaultAcronyms` listed `"Bom"`/`"BoM"` and `"Uom"`/`"UoM"` as separate entries, but the underlying `Dictionary<string, string>` collapsed each pair to a single last-write-wins survivor. This caused two `AcronymRegistryTests` to fail on the PR head (`Expected: "Bom" But was: "BoM"`).

The registry now stores an ordered list of registered variants per upper-invariant key:

- Storage changed from `Dictionary<string, string>` to `Dictionary<string, List<string>>`.
- New public API: `TryGetVariants(word, out IReadOnlyList<string> variants)`.
- `TryGetCanonical` unchanged in signature; now returns the **first-added** variant for the key.
- User entries **displace** built-in variants for their key (preserves existing override semantics); multiple user entries under the same key accumulate in order.
- `DefaultAcronyms` reordered so the canonical variant precedes additional variants (`"BoM", "Bom"`, `"UoM", "Uom"`).

`IdentifierNameRenderer.GetWordAlternatives` (`HasAnyUpper` branch) now consumes `TryGetVariants` and appends every registered variant that differs from the primary. Result: subscribers can list multiple acceptable spellings for the same acronym (e.g. `BoM` + `Bom` both accepted alongside the source's own casing). The all-lowercase branch is unchanged.

Config and defaults remain flat arrays — no schema change.

### 2. Collision guard: case-insensitive AL identifier comparison

`WouldCollideInContainingType` compared sibling names with `StringComparison.Ordinal`. AL treats duplicate method identifiers case-insensitively (`AL0018`), so an only-case-different sibling (`mypublisher_onsomething` vs. suggested `MyPublisher_OnSomething`) was not detected and the CodeFix would produce a duplicate-identifier compile error.

Both sibling comparisons now use `SemanticFacts.IsSameName` (repo convention per `analyzer-development.instructions.md`). The rule-enforcement comparison at the top of `Analyze` (accepted-set membership) intentionally stays `StringComparison.Ordinal` — enforcing exact casing is the rule's purpose.

Regression fixture: `NoDiagnostic/PreferredNameCollidesWithSiblingDifferentCase.al`.

### 3. Resx: default template placeholder spelling

`EventSubscriberNamingPatternDescription` documented the default template as `'{Event Source}_{EventName}[{Element Name}]'`. Actual default (and schema description) is `'{Event Source}_{Event Name}[_{Element Name}]'`. Placeholder spelling is semantically significant (raw vs PascalCase), and the `_` inside the optional group matters. Corrected.

## Test results

- `ALCops.Common.Test`: 68/68 passing (was 66/68 failing on PR head; +4 new multi-variant tests, −2 removed per-entry assertions replaced).
- `ALCops.LinterCop.Test` (EventSubscriberNamingPattern filter): 34/34 passing (+1 regression case).

## Compatibility

- Public API additive: new `TryGetVariants`; `TryGetCanonical` signature and semantics unchanged for keys with a single variant.
- `alcops.json` `KnownAcronyms` schema unchanged (flat array).
- LC0098 accepts a strict superset of previous spellings when defaults or user config list multiple variants; no previously-accepted spelling is now rejected.

Arthurvdv (approved):

Arthurvdv:
Thans for looking into the feedback. Let's get this merged :-)

PR #410, opened 2026-07-17
fix(lc0083): suppress false positive on FieldRef.Value

### LINKED ISSUE #409: [Bug]: LC0083 false positive on FieldRef.Value in DT2Date/DT2Time/Format/Date2DMY/Date2DWY
### Analyzer

LinterCop

### Issue type

### Rule ID

LC0083

### Description

LC0083 (BuiltInDateTimeMethod) incorrectly reports a diagnostic when a built-in date/time method is called with FieldRef.Value as the argument. The suggested replacement (e.g. MyFieldRef.Value.Date()) is invalid because FieldRef.Value returns the AL wildcard/dynamic type Joker in the SDK, which does not expose the new .Date(), .Time(), .Day(), etc. extension methods.

### Root cause

- FieldRef.Value is modeled by the AL SDK as an IInvocationExpression (getter call) whose return type is NavTypeKind.Joker, not NavTypeKind.Variant.
- The existing IsVariantArgument guard only checked for Variant and therefore missed Joker, so the analyzer proceeded to emit a diagnostic.

### Fix

- Extend the variant guard to also treat NavTypeKind.Joker as a skip condition (renamed to IsVariantOrJokerArgument).
- Add a dedicated IsFieldRefValueAccess guard that detects .Value accesses on a FieldRef receiver, covering both IInvocationExpression (current SDK model) and IFieldAccess (defensive against future SDK changes).
- Restrict the FieldRef guard to the Value member only (via case-insensitive IsSameName), so other FieldRef members (.Name, .GetFilter(), …) remain unaffected.

### AL code to reproduce

```al
table 50100 MyTable
{
    fields
    {
        field(1; "My DateTime"; DateTime) { }
    }
}

codeunit 50100 MyCodeunit
{
    procedure MyProcedure()
    var
        MyTable: Record MyTable;
        MyRecRef: RecordRef;
        MyFieldRef: FieldRef;
        MyDate: Date;
    begin
        MyRecRef.GetTable(MyTable);
        MyFieldRef := MyRecRef.Field(1);
        MyDate := DT2Date(MyFieldRef.Value); // LC0083 reported, but replacement is invalid
    end;
}
```

MODUSCarstenScholling:
FieldRef.Value returns NavTypeKind.Joker (AL wildcard), not Variant, so IsVariantArgument missed it and the analyzer emitted an invalid replacement suggestion like `MyFieldRef.Value.Date()`.

- Extend the variant guard to also treat Joker as a skip (renamed to IsVariantOrJokerArgument).
- Add IsFieldRefValueAccess that matches `.Value` on a FieldRef receiver via IInvocationExpression (current SDK) or IFieldAccess (defensive).
- Restrict the FieldRef guard to the `Value` member using IsSameName so other FieldRef members are unaffected.
- Add fixtures: RecRefFieldRefDT2DateTime, FieldRefValueChained, FormatFieldRefValue (NoDiagnostic) and DT2XOnRecordField (HasDiagnostic).

Fixes #409

### DIFF
--- a/.github/instructions/instruction-maintenance.instructions.md
+++ b/.github/instructions/instruction-maintenance.instructions.md
@@ -96,6 +96,7 @@ Include: project structure, templates, step-by-step guides, API reference, commo
 | `fc0004-permission-declaration-order` | rule-scoped | FC0004 rule |
 | `fc0005-use-parenthesis-for-method-assignment` | rule-scoped | FC0005 rule |
 | `fc0006-permission-values-should-be-lowercase` | rule-scoped | FC0006 rule |
+| `lc0083-built-in-date-time-method` | rule-scoped | LC0083 rule |
 | `lc0086-page-style-string-literal` | rule-scoped | LC0086 rule |
 | `lc0091-translatable-text-should-be-translated` | rule-scoped | LC0091 rule |
 | `lc0092-naming-pattern` | rule-scoped | LC0092 rule |
--- a/.github/instructions/lc0083-built-in-date-time-method.instructions.md
+++ b/.github/instructions/lc0083-built-in-date-time-method.instructions.md
@@ -1,0 +1,57 @@
+---
+applyTo: 'src/ALCops.LinterCop/**/BuiltInDateTimeMethod*'
+---
+
+# LC0083: BuiltInDateTimeMethod
+
+## Purpose
+
+Detects calls to outdated built-in date/time functions (`Date2DMY`, `Date2DWY`, `DT2Date`, `DT2Time`, `Format(... , 0, '<HOURS24|MINUTES|SECONDS|THOUSANDS>')`) and suggests the modern extension methods on `Date`, `Time`, and `DateTime` values (`.Date()`, `.Time()`, `.Day()`, `.Month()`, `.Year()`, `.DayOfWeek()`, `.WeekNo()`, `.Hour()`, `.Minute()`, `.Second()`, `.Millisecond()`). Ships with `BuiltInDateTimeMethodCodeFixProvider`.
+
+## Diagnostic properties
+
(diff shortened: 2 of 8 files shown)


Arthurvdv (approved):
**Review: Approve** ✅

Verified the fix end-to-end: built all three TFMs (net10.0/net8.0/netstandard2.1) with 0 warnings, ran all 28 `BuiltInDateTimeMethod` tests (all pass), and confirmed the three new NoDiagnostic fixtures fail against the `main` analyzer but pass with this fix — they genuinely reproduce #409.

**Correctness**
- No new false negatives: `IsFieldReferenceValueAccess` requires both the `Value` member name (nullable-safe `IsSameName`) *and* a FieldRef-typed receiver, so record fields named `Value` and other FieldRef members (`.Name`, `.Number`) are unaffected. `DT2XOnRecordField` guards against over-suppression.
- Suppressing on Joker is correct: the `.Date()`/`.Hour()` extension methods don't bind on Joker-typed expressions, mirroring the existing Variant rationale.

**Observation (not blocking)**
- `IsFieldReferenceValueAccess` is currently fully shadowed by the Joker check, since `FieldRef.Value` types as Joker. It's harmless (one pattern match, no extra semantic-model calls) and documented as deliberate defense against SDK remodeling.

**Conventions**
- `EnumProvider.NavTypeKind.*` used throughout, no new `GetSymbol()` calls, `ctx.IsObsolete()` intact, guards run before replacement-syntax construction (cheap-first ordering preserved).
- Instruction file test counts (9 HasDiagnostic / 8 NoDiagnostic / 11 HasFix) match fixtures on disk; the `instruction-maintenance` change is just the required index row.

---

Arthurvdv:
Thanks for catching and fixing this bug!

PR #412, opened 2026-07-17
refactor(TranslationIdHelper): encapsulate translation-ID casing workaround

### LINKED ISSUE #411: [Bug]: LC0091 False-positive missing translation when property casing in source differs from canonical
### Analyzer

LinterCop

### Issue type

### Rule ID

LC0091

### Description

`TranslatableTextShouldBeTranslated` (LC0091) reports a false-positive missing translation when the developer writes a translatable property using non-canonical casing (e.g. `Tooltip` or `TOOLTIP` instead of `ToolTip`). The XLIFF file produced by the AL compiler contains the correct translation, but the analyzer computes a different trans-unit ID and therefore fails to find it.

XLIFF (compiler-generated, canonical casing):

```xml
<trans-unit id="Page 2931038265 - Control 1296262074 - Property 1295455071" ...>
  <source>This is a tooltip.</source>
  <target>Dies ist ein ToolTip.</target>
  <note ...>Page MyPage - Control MyField - Property ToolTip</note>
</trans-unit>
```

Analyzer output:

```
LC0091: Missing translation for 'Tooltip' in language(s): de-DE
```

Expected: no diagnostic — the translation exists in the XLIFF.

### Root cause

The trans-unit ID is composed of hashed segments (object, control, property). The AL compiler hashes the **canonical** property name (`ToolTip` → `1295455071`), while the analyzer hashes the **source-cased** name (`Tooltip` → `2001309823`). The IDs do not match, so the translation lookup fails.

The SDK's translation-ID methods (`LanguageFileUtilities.GetTranslationFileId` on newer SDKs, `GetLanguageSymbolId` / `GetLabelTextConstLanguageSymbolId` on older ones) derive the hashed segment from `symbol.Name` on the live `IPropertySymbol` instance. There is no public API to override the name, and the `name` string parameter on `GetTranslationFileId` only feeds the human-readable note segment — not the hash.

### Impact

- Applies to any translatable property (`Caption`, `ToolTip`, `PromotedActionCategories`, `InstructionalText`, `AboutTitle`, `AboutText`, `OptionCaption`, …) written with non-canonical casing.
- AL is case-insensitive, so the source compiles cleanly; only LC0091 reports a false positive.
- Affects both new-SDK (`GetTranslationFileId`) and old-SDK (`GetLanguageSymbolId`) code paths in the helper.

### AL code to reproduce

Source (`Tooltip` with lowercase `t`):

```al
page 50100 MyPage
{
    layout
    {
        area(Content)
        {
            field(MyField; MyField)
            {
                Tooltip = 'This is a tooltip.';
            }
        }
    }
}
```

MODUSCarstenScholling:
Move the private `name`-field mutation for canonical property casing out of the LC0091 analyzer and into `TranslationIdHelper.ComputeTranslationId` via a new optional `nameOverride` parameter.

- Explicit `GetField("name")` + string-type guard (was: fragile Contains match)
- Original value restored in `finally` (was: permanent mutation)
- Applies to both new-SDK and old-SDK translation-ID paths
- Cached FieldInfo per symbol type (ConcurrentDictionary)
- Per-symbol lock via ConditionalWeakTable serializes concurrent mutations

Add regression test `NoDiagnosticPropertyCasing / PageControlToolTipWrongCasing` covering a page control with `Tooltip` (source) matched against an XLIFF trans-unit id built from `ToolTip` (canonical).

Fixes #411

### DIFF
--- a/.github/instructions/lc0091-translatable-text-should-be-translated.instructions.md
+++ b/.github/instructions/lc0091-translatable-text-should-be-translated.instructions.md
@@ -30,6 +30,7 @@ These decisions were made during the initial design and should be preserved unle
 | Severity | Warning | Missing translations directly affect user experience |
 | Extension root symbol | `ExtensionObjectFoldingUtilities.GetTranslationRootSymbol` SDK API | Fixes the multi-extension bug in the original rule; correctly handles all cases |
 | Translation ID generation | `LanguageFileUtilities` via **runtime reflection** (`GetTranslationFileId` on new SDKs, public 2-param `GetLanguageSymbolId` / `GetLabelTextConstLanguageSymbolId` on older SDKs) | The public methods gained an optional `useNamespaces` param in `18.0.38.52553`; C# bakes optional-param call sites at compile time, so a direct call breaks across SDK versions (see Known issues). Reflection picks the right method at runtime and supports namespace-aware IDs. |
+| Canonical property-name override | Passed only when `manifest.Runtime > RuntimeVersion.Spring2020CU1` (or runtime unknown → treat as "current") | Mirrors `SymbolExtensions.GetTranslationName`: the compiler uses `PropertyKind.ToString()` on new runtimes but source-cased `symbol.Name` on runtime ≤ 5.1. Unconditional override regresses legacy apps (source-cased hash ≠ canonical hash → false positive). Gate is applied on the LinterCop side; `TranslationIdHelper.ComputeTranslationId(nameOverride: null)` falls through to `symbol.Name`. |
 | ManifestHelper exception handling | Catch `FileNotFoundException` and treat as null manifest | `ManifestHelper.GetManifest` loads `Microsoft.Dynamics.Nav.Analyzers.Common` assembly via reflection; this assembly isn't present in test contexts |
 | Null manifest behavior | Proceed with analysis (don't skip) | Tests create minimal compilations without manifests; real projects always have manifests |
 | XLIFF caching | Load once in CompilationStartAction | Avoid re-parsing per symbol |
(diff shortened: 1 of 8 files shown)


Arthurvdv (changes requested):
## Code review

### 🔴 High: The symbol mutation is unnecessary on the primary SDK path (and the doc justifying it is wrong)

**`src/ALCops.Common/Reflection/TranslationIdHelper.cs:121–176`** — The XML doc asserts the SDK's `GetTranslationFileId` hashes the property/label name from the live symbol instance rather than from the `name` string parameter. Verified against the decompiled SDK source (net8.0 and net10.0): this is **false**. `GetTranslationFileId(name, kind, containingSymbol, ...)` builds the leaf trans-unit segment from the **string parameter** (`LanguageFileUtilities.GetLanguageSymbolId`, line 120: `instance.Add(GetSingleSymbolId(name, kind, shouldHashName, useNamespaces));`) and never reads the leaf symbol at all. The current code only works because the lambda re-reads `symbol.Name` *inside* the mutation window, so the override ends up in the parameter anyway.

Consequence: on the dominant new-SDK path, the code takes a per-symbol lock and temporarily rewrites the private `name` field of a **shared, compiler-owned symbol** for zero benefit. During that window, any other thread (other analyzers' callbacks on the same compilation — ConcurrentExecution is enabled — or SDK code walking members) that reads `property.Name` lock-free observes the wrong (canonical) name. The lock only serializes this helper against itself.

**Suggested fix:** On the new-SDK path, pass `nameOverride ?? symbol.Name` as the first `Invoke` argument and drop the mutation there entirely. The mutation (or better: reflectively binding the internal string-based `GetLanguageSymbolId(string, SymbolKind, Symbol, bool, ...)` overload, which also exists on old SDKs) is only needed for the old-SDK fallback. Fix the doc comment to match.

### 🟡 Medium: Unconditional canonical-name override regresses apps with runtime ≤ 5.1 (BC16 Spring 2020 CU1 and older)

**`src/ALCops.LinterCop/Analyzers/TranslatableTextShouldBeTranslated.cs:225`** (`nameOverride: propertyKind.ToString()`) — The compiler does **not** always use the canonical property name for XLIFF IDs. From the decompiled SDK (`SymbolExtensions.GetTranslationName`):

```csharp
if (!(runtimeVersion <= RuntimeVersion.Spring2020CU1))   // Spring2020CU1 = 5.1
    return property.GetPropertyTypeInfo().Kind.ToString();
return property.Name;   // source-cased!
```

For an app with runtime ≤ 5.1, the compiler hashes the **source-cased** name (`Tooltip` → 2001309823), but the analyzer now unconditionally hashes the canonical name (`ToolTip` → 1295455071). Result: exactly the false positive #411 fixes, reintroduced for old-runtime apps — where the pre-PR code was correct. The repo explicitly supports BC16 (`RequireMinimumVersion("16.0")` in tests).

**Suggested fix:** Mirror the compiler's gate — only apply the canonical override when the declaration's runtime version is > 5.1 (`LabelWriterVisitor.VisitTranslatableProperty` uses `p.GetTranslationName(p.DeclaringSyntaxNode.GetRuntimeVersionOrCurrent())`).

### 🟡 Medium (policy): Instruction file not updated

`.github/instructions/lc0091-translatable-text-should-be-translated.instructions.md` — The "Known issues and workarounds" section was not updated for this PR's design decision (canonical-name override, private `name`-field mutation, per-symbol locking). Repo policy requires instruction files to stay in sync with such workarounds. Please document the casing workaround and its runtime-version caveat.

---

**Verified non-issues (no action needed):**
- The regression test is genuine: recomputed the SDK's FNV-1a hashes — `Tooltip`=2001309823 ≠ `ToolTip`=1295455071 — so without the fix the first marker produces a diagnostic and the test fails; the second control guards the baseline.
- `GetField("name")` works: the field is declared directly on `SourcePropertySymbol`, instance `initonly` fields are settable via reflection on .NET 8/10, and `Name => name` reads the field directly. String-type guard and null-field fallback degrade gracefully per repo convention.
- netstandard2.1: the file remains inside `#if !NETSTANDARD2_1`, so `ConditionalWeakTable`/`ConcurrentDictionary` API concerns don't apply.
- Report-label and label-const paths correctly omit `nameOverride` (those XLIFF IDs use the user-chosen source name).
- The copied `SkipTestIfVersionIsTooLow` block in the new test is dead code but harmless.

---

MODUSCarstenScholling:
## LC0091 canonical property-name override — runtime-version gate + tests

**Verdict: LGTM, ready to merge.** Solution-wide tests: **1284 pass / 0 fail / 2 pre-existing skips**.

**1. HIGH — Symbol mutation in `TranslationIdHelper.cs` ✅**
- New-SDK path (`GetTranslationFileId`) now passes `effectiveName = nameOverride ?? symbol.Name` as the SDK's `name` string parameter — **no** private-field mutation and **no** per-symbol lock on the hot path.
- Old-SDK fallback (public 2-param `GetLanguageSymbolId` / `GetLabelTextConstLanguageSymbolId`) still needs `InvokeWithCanonicalName` because those overloads read `symbol.Name` internally with no seam. Mutation is bounded to that call and released in `finally`.
- XML docs cleanly separate the two paths; speculative wording about impact scope removed.

**2. MEDIUM — Unconditional canonical override regresses runtime ≤ 5.1 ✅**
- `useCanonicalPropertyName = manifest?.Runtime is null || manifest.Runtime > RuntimeVersion.Spring2020CU1` computed once in `OnCompilationStart`, threaded through `AnalyzeSymbol` → `AnalyzePageLikeSymbol` → `ReportTranslatableProperty`.
- On legacy runtimes: `nameOverride = null` → SDK hashes source-cased `symbol.Name`, matching `SymbolExtensions.GetTranslationName`.
- Unknown runtime defaults to "current" (mirrors the SDK's `GetRuntimeVersionOrCurrent`).
- **Regression coverage:** paired `HasDiagnosticLegacyRuntime` (canonical XLIFF only → diagnostic expected) and `NoDiagnosticLegacyRuntime` (source-cased XLIFF only → no diagnostic). Both would fail under the old unconditional-override behavior.

**3. MEDIUM (policy) — Instruction file updated ✅**
- Design-decisions table: new "Canonical property-name override" row.
- New "Known issues" subsection covering the runtime-version gate + regression-test setup (`features: ["TranslationFile"]` + `Nav.Analyzers.Common.dll` in test bin).
- New "Canonical-name override mechanism" subsection covering new-SDK string-parameter path, old-SDK private-field mutation, `_nameFieldCache`, `_symbolLocks` (`ConditionalWeakTable`), and caveats.
- Test-coverage summary updated (`HasDiagnosticLegacyRuntime`, `NoDiagnosticLegacyRuntime`, `NoDiagnosticPropertyCasing`; stale `NoDiagnostic` count corrected 2 → 3).

### Notes

- The public 2-param SDK overloads would ideally go away: if Microsoft promoted `LanguageFileUtilities.GetTranslationFileId` from `internal` to `public`, the reflection cache, the private-field mutation, and the per-symbol lock could all be deleted. Worth an upstream `microsoft/AL` API-suggestion, but out of scope for this PR.

Arthurvdv (approved):

Arthurvdv:
I agree with the LGTM, so let's get this merged.

PR #415, opened 2026-07-20
fix(multiple) missing MessageFormat arguments across analyzers

### LINKED ISSUE #414: [Bug]: Several analyzers have diagnostic messages containing incorrect placeholders
## Symptoms

A number of `Diagnostic.Create` calls pass fewer or more arguments than their
`MessageFormat` string references, so the rendered diagnostic message
still contains the unformatted `{n}` placeholder instead of
the expected symbol name.

## Affected rules

| ID | Rule | Effect |
|---|---|---|
| `PC0001` | EditableFlowField | Message shows `{0}` instead of the field name |
| `PC0011` | EventPublisherIsHandledByVar | Message shows `{0}` instead of the parameter name |
| `LC0054` | InterfaceObjectNameGuide | Message shows `{0}` instead of the interface name |
| `LC0082` | UseQueryOrFindWithNextInsteadOfCount | `Count() >= 1` is reported as `Count() <>` - operator collapsed to `<>`, compared value omitted |
| `AC0001` | LookupPageIdAndDrillDownPageId | Second argument passed but never rendered - list page name missing from message |
| `AC0012` | IntegrationEventInInternalCodeunit | Extra argument passed that the message doesn't reference |

## Cause

Two independent classes of issues:

1. **Missing args at the call site** - `Diagnostic.Create(descriptor, location)` is called with no message args although the `MessageFormat` contains `{0}`. Affects `PC0001`, `PC0011`, `LC0054`.
2. **Resx / call-site drift** - the number of args passed and the number of placeholders in the resx no longer match, either because the message was later reworded or the call site was later extended. Affects `AC0001`, `AC0012`.

Additionally, `LC0082`'s helper `GetOperatorKind` normalized every
non-`=` operator to `<>` and threw away the compared literal value, so
the emitted message could not describe the actual code being flagged.

## Expected behaviour

Every diagnostic message renders with all placeholders substituted; the
operator and compared value in `LC0082`'s message match the source code
being flagged.

MODUSCarstenScholling:
Several diagnostics passed too few arguments to Diagnostic.Create, producing messages with literal "{0}" placeholders. Add the missing name arguments and align resx/call sites where they had drifted.

- AC0001 (LookupPageIdAndDrillDownPageId): include list page name in message
- AC0012 (IntegrationEventInInternalCodeunit): drop superfluous {1} arg
- PC0001 (EditableFlowField): pass field.Name
- PC0011 (EventPublisherIsHandledByVar): pass parameter.Name
- PC0013: diagnostic message now includes the resolved primary key fields for clarity.
- LC0054 (InterfaceObjectNameGuide): pass interfaceTypeSymbol.Name (3 sites)
- LC0082 (UseQueryOrFindWithNextInsteadOfCount): preserve real operator and compared value; add FlipOperator/GetOperatorSign switch expressions so "2 > Rec.Count()" is normalized to "Count() < 2" in the message

Fixes #414

### DIFF
--- a/src/ALCops.ApplicationCop/ALCops.ApplicationCopAnalyzers.resx
+++ b/src/ALCops.ApplicationCop/ALCops.ApplicationCopAnalyzers.resx
@@ -277,7 +277,7 @@
     <value>DrillDownPageId and LookupPageId property must be defined for tables used in list pages</value>
   </data>
   <data name="LookupPageIdAndDrillDownPageIdMessageFormat" xml:space="preserve">
-    <value>Table '{0}' is used in a list page but does not define both the DrillDownPageId and LookupPageId properties.</value>
+    <value>Table '{0}' is used in list page '{1}' but does not define both the DrillDownPageId and LookupPageId properties.</value>
   </data>
   <data name="LookupPageIdAndDrillDownPageIdDescription" xml:space="preserve">
     <value>When a table is used as the source for a list page, users expect to be able to drill down into records and open lookups in a consistent and predictable way. The DrillDownPageId and LookupPageId table properties define which pages are used for these interactions.</value>
--- a/src/ALCops.ApplicationCop/Analyzers/IntegrationEventInInternalCodeunit.cs
+++ b/src/ALCops.ApplicationCop/Analyzers/IntegrationEventInInternalCodeunit.cs
@@ -38,8 +38,7 @@ public sealed class IntegrationEventInInternalCodeunit : DiagnosticAnalyzer
         ctx.ReportDiagnostic(Diagnostic.Create(
             DiagnosticDescriptors.IntegrationEventInInternalCodeunit,
             methodSymbol.GetLocation(),
-            methodSymbol.Name,
-            applicationObject.Name));
+            methodSymbol.Name));
     }
 
     private static bool IsIntegrationEvent(IMethodSymbol methodSymbol) =>
--- a/src/ALCops.LinterCop/ALCops.LinterCopAnalyzers.resx
+++ b/src/ALCops.LinterCop/ALCops.LinterCopAnalyzers.resx
@@ -355,7 +355,7 @@
     <value>Consider using a Query object or Rec.Find('-') with Rec.Next() for checking exactly one record</value>
   </data>
   <data name="UseQueryOrFindWithNextInsteadOfCountMessageFormat" xml:space="preserve">
-    <value>Consider using a Query object or {0}.Find('-') together with {0}.Next() instead of {0}.Count() for faster and more efficient record checks.</value>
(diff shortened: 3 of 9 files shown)


MODUSCarstenScholling:
The change for PC0013 is based on internal feedback.

Arthurvdv (approved):
## Review: PR #415 — fix(multiple) missing MessageFormat arguments across analyzers

**Summary:** Verified every touched rule end-to-end: resx placeholder count/order against all `Diagnostic.Create` call sites per descriptor, the LC0082 operator-normalization logic, PC0013 TFM guards/reflection helpers, and test impact.

- **AC0001** — resx now has `{0}`/`{1}`; the single call site already passed `table.Name` + the page name in matching order. ✓
- **AC0012** — the dropped `applicationObject.Name` arg was superfluous; `applicationObject` is still used for the internal-codeunit guard, so no unused-variable fallout. ✓
- **PC0001 / PC0011** — one call site each, now passing the single `{0}` arg. ✓
- **LC0054** — all 3 call sites updated consistently; no other call sites exist. ✓
- **PC0013** — new resx format has 3 placeholders; both call sites pass `GetPrimaryKeySignature(table)`, `table.Name`, `expectedArgs` in matching order. `GetMethodName` is a `const string`, valid in the interpolation. `QuoteIdentifierIfNeededWithReflection` and `ToDisplayStringWithReflection` exist in ALCops.Common with matching `#if NETSTANDARD2_1` guards; the `#else` path (`field.Type?.ToDisplayString()`) mirrors existing usage. All three TFMs (netstandard2.1/net8/net10) check out. ✓
- **LC0082** — exactly one side of the comparison can be an int literal (the `Count()` side yields the −1 sentinel), so `countIsOnRight = leftValue >= 0 && rightValue < 0` is sound. Every report path guarantees `comparedValue` is 1 or 2, making the `"="` fallback in `GetOperatorSign` unreachable. `FlipOperator` mirrors correctly: `2 > Rec.Count()` → "Count() < 2", `1 <= Rec.Count()` → "Count() >= 1". A genuine `-1` literal can't collide with the sentinel since only 0/1/2 trigger reports, and 0-comparisons are diverted to the IsEmpty diagnostic earlier. ✓
- **Tests** — all touched rules use marker-based assertions (`HasDiagnosticAtAllMarkers`) with diagnostic IDs only; no test asserts message text, and arg counts now match placeholders, so no `FormatException` risk at render time.

### Findings

**1. Missing trailing newline**
**File:** `src/ALCops.PlatformCop/Analyzers/RecordGetProcedureArguments.cs` (last line)
**Severity:** Trivial
No functional impact.

### Verdict

**Approve.** No bugs, logic errors, or placeholder/arg mismatches found — the PR does what it claims at every call site, and the LC0082 operator normalization is correct across all reachable paths.

Issue #423, opened 2026-07-29
[Feature]: FC0007 Statement blocks should be separated by a blank line,


MODUSCarstenScholling:
### Motivation

AL code frequently mixes multi-line control-flow blocks (`if … then begin … end`,
`case`, `repeat`, `while`, `for`, `foreach`) with adjacent statements without a
visual break. Same for scope-leaving statements (`exit`, `Error()`) and
`if … end else … end` chains. The result is dense, hard-to-scan procedures where
the boundary between a control block and the code around it is invisible.

Example (accepted today, flagged after this change with defaults):

```al
Foo := 1;
if X then begin
    DoSomething();
end;
Bar := 2;
```

Expected:

```al
Foo := 1;

if X then begin
    DoSomething();
end;

Bar := 2;
```

### Proposal

Introduce **FC0007 — Statement blocks should be separated by a blank line**,
shipped in `ALCops.FormattingCop`. Opt-in (disabled by default), configurable
via `alcops.json` under a new `StatementBlockSpacing` object.

Configuration axes:

| Property | Type | Default | Purpose |
|---|---|---|---|
| `ControlFlowBefore` | `bool` | `true` | Require blank line before `if/case/…` |
| `ControlFlowAfter` | `bool` | `true` | Require blank line after the terminating `end;` |
| `ScopeLeavingMode` | enum | `ExitAndError` | Which scope-leaving statements to check (`Off` / `ExitOnly` / `ErrorOnly` / `ExitAndError`) |
| `ElseChainBeforeMode` | enum | `Off` | Whether `end else` requires a blank line (`Off` / `RequireBlank`) |
| `OneLinerMode` | enum | `Off` | Whether single-line control-flow statements participate (`Off` / `All`) |

### Non-goals for the first cut

- No CodeFix (spacing edits are cheap manually; a CodeFix requires care with
  comment attachment).
- No `CaseBranchMode`, `GuardClauseMode`, `LoopControlBeforeMode`,
  `TreatCommentOnlyLinesAsSeparator`, `SkipDirectiveBoundaries` — captured on
  the FC0007 roadmap in the rule instruction file.

### Acceptance

- Rule disabled by default in `DiagnosticDescriptors`.
- Every property has default behavior identical to the "no config file" case.
- Malformed `alcops.json` (unknown enum values, syntax errors) must not throw
  from the analyzer — provider should silently fall back to defaults.
[...]

Issue #328, opened 2026-06-07
Emit diagnostic warning when alcops.json is found but cannot be parsed


Arthurvdv:
## Context

When `alcops.json` is found in the file hierarchy but cannot be read or parsed (malformed JSON, permission errors on the file itself), the settings provider currently returns default settings silently.

This can be confusing for users who have a typo in their config file and don't understand why their settings aren't being applied.

## Proposed behavior

Emit a diagnostic warning (not an error) when:
- `alcops.json` is found but cannot be deserialized (malformed JSON)
- `alcops.json` is found but cannot be read (file-level permission error)

The warning should indicate the file path and the reason for failure.

## Related

- #322 (parent directory traversal for settings lookup)
- #325 (settings provider overload consolidation)

PR #424, opened 2026-07-29
feat(FC0007): Statement blocks should be separated by a blank line (opt-in)

MODUSCarstenScholling:
## feat(FC0007): Statement blocks should be separated by a blank line (opt-in)

Adds **FC0007 — Statement blocks should be separated by a blank line** to
`ALCops.FormattingCop`. Opt-in (disabled by default) and fully configurable via
`alcops.json`. Closes #<issue-number>.

### What it flags

- Multi-line control-flow blocks (`if`, `case`, `repeat`, `while`, `for`,
  `foreach`) without a blank line before them and/or after their terminating
  `end;`.
- Scope-leaving statements: `exit`, `exit(<expr>)`, and built-in `Error(...)`
  calls that are not preceded by a blank line.
- Optionally the `end else` transition in `if … end else … end` chains.

Single-line (one-liner) control-flow statements are ignored unless
`OneLinerMode` is set to `All`.

### Configuration surface

New object in `alcops.json`:

| Property | Type | Default |
|---|---|---|
| `ControlFlowBefore` | `bool` | `true` |
| `ControlFlowAfter` | `bool` | `true` |
| `ScopeLeavingMode` | `Off` / `ExitOnly` / `ErrorOnly` / `ExitAndError` | `ExitAndError` |
| `ElseChainBeforeMode` | `Off` / `RequireBlank` | `Off` |
| `OneLinerMode` | `Off` / `All` | `Off` |

Enums are real C# types (`ALCops.Common.Settings.{Scope,ElseChainBefore,OneLiner}Mode`),
serialized case-insensitively via `JsonStringEnumConverter` on `net8+` and
`StringEnumConverter` on `netstandard2.1`. Both converter registrations were
added to `ALCopsSettingsProvider`.

### Provider hardening (small, but notable)

`ALCopsSettingsProvider.DeserializeSettings` now swallows `JsonException` and
returns defaults. Users with a broken `alcops.json` (e.g. an unknown enum
value) previously would have crashed the analyzer callback and produced an
`AD0001`; now the rule simply behaves as if no config file existed. A dedicated
test guards this behavior with a fixture containing `"ScopeLeavingMode":"NotAnEnumValue"`.

### Schema parity guard

`Settings/alcops.schema.json` mirrors the three FC0007 enums. Any drift is
caught by `StatementBlockSpacingSchema.cs`, a NUnit test template that reflects
over the C# enum values and compares them against the `enum` arrays in the
schema JSON. Reusable pattern for future enum-typed settings.

### Test coverage

- 18 rule tests (positive/negative × control-flow / scope-leaving / else-chain /
  one-liner / disabled-by-default / malformed-config regression).
- 3 schema-parity tests (one per enum).
- 21/21 passing; full solution 1407 tests remain green.

Tests follow a single `[Test] + [TestCase]` pattern across two parametrized
methods (`HasDiagnostic` / `NoDiagnostic`) driven by a named settings
dictionary. The fixture folder (`HasDiagnostic/` vs `NoDiagnostic/`) is
resolved implicitly by `LoadFixtureAsync` — test signatures stay free of
storage-layout details.

### Docs

- New rule instruction file
  `.github/instructions/fc0007-statement-blocks-separated-by-blank-line.instructions.md`
  with design decisions, test summary counts, known issues, and a Phase-2
  roadmap (`CaseBranchMode`, `GuardClauseMode`, `LoopControlBeforeMode`,
  `TreatCommentOnlyLinesAsSeparator`, `SkipDirectiveBoundaries`).
- `common-library.instructions.md` updated for the new `StatementBlockSpacing`
  setting and the "add a new setting" enum-converter walkthrough.
- Site page will be added in a follow-up to
  [alcops.dev](https://github.com/ALCops/alcops.dev)
  at `content/docs/analyzers/formattingcop/fc0007.md`.

### Not in this PR (roadmap)

- CodeFix (comment-attachment edge cases need care).
- Additional spacing axes listed above.

Implements #423


Arthurvdv:
## Code Review: FC0007 + bundled LC0095 rewrite

Reviewed against `origin/main...origin/pr-424` (merge-base `7932e20`). All three TFMs (netstandard2.1/net8.0/net10.0) build clean and the new test suites pass (18 FC0007 + 25 LC0095). Findings ordered by severity:

### :red_circle: High — `exit`/`Error()` in an `else` branch produces false positives

`src/ALCops.FormattingCop/Analyzers/StatementBlocksSeparatedByBlankLine.cs` (`AnalyzeExitStatement` ~line 163, `AnalyzeInvocationExpression` ~line 204)

`AnalyzeControlFlowStatement` deliberately skips statements whose `Parent is IfStatementSyntax` (line 73), but the `exit`/`Error()` paths have no such guard. `GetSiblingStatements` on an `IfStatementSyntax` returns `[thenStatement, elseStatement]`, so an `exit` in the else branch gets index 1 and is checked for a blank line against the then-branch:

```al
if Flag then begin
    DoSomething();
end else
    exit;                                  // FP: "before scope-leaving statement 'exit'"

if Flag then DoSomething() else exit;      // FP: same-line else branch, diff = 0
```

No fixtures cover an else-branch `exit`/`Error()`. **Fix:** add the same `Parent is IfStatementSyntax` early-return to both methods, plus fixtures for `else exit;` / `end else Error(...)`.

### :yellow_circle: Medium — Comment-only lines count as blank lines, contradicting the rule''s own spec

`HasBlankLineBetween` (~line 357) only compares token line numbers (`diff >= 2`). Comments and `#region`/`#pragma` directives are trivia, not tokens, so:

```al
Message(''Start'');
// ---- section divider ----
exit;    // NOT flagged: line diff = 2, but no actual blank line
```

The bundled instruction file states the opposite ("only actual blank lines count"). Either the doc or the implementation is wrong. **Fix:** decide intended semantics; if comments must not count, inspect trivia between tokens for whitespace-only lines, and add a comment fixture.

### :yellow_circle: Medium — `"StatementBlockSpacing": null` in alcops.json crashes the analyzer

`ALCopsSettings.StatementBlockSpacing` is non-nullable with a `new()` default, but neither STJ nor Newtonsoft enforces NRT annotations. An explicit `null` deserializes without `JsonException` (so the new try/catch never triggers), gets cached per directory, and every subsequent `config.ControlFlowBefore` access throws NRE — violating the "malformed alcops.json → defaults" contract this PR adds (#328). **Fix:** null-coalesce at consumption (`settings.StatementBlockSpacing ?? new()`) or normalize after deserialization.

### :green_circle: Low — LC0095 FixAll span-scope branch is a latent no-op

`ParameterNotReferencedCodeFixProvider.cs` (`FixAllAsync`): when `fixAllSpans` is present, each span covers a whole member; `root.FindNode(span)` + `AncestorsAndSelf().OfType<ParameterSyntax>()` resolves the member declaration, never a parameter, so nothing is removed. Currently unreachable (parameterless `FixAllProvider.Create` only exposes Document/Project/Solution scopes), but the code claims support. **Fix:** intersect `GetDocumentDiagnosticsAsync` results with the spans and use diagnostic locations for `FindNode`.

### Minor

- `StatementBlocksSeparatedByBlankLineDescription` (resx) promises "case branches should be separated from each other", but `CaseBranchMode` is roadmap-only and `DisabledByDefault.al` asserts case branches are *not* flagged. User-facing metadata should match shipped behavior.
- `EnumProvider.SyntaxKind.CaseElse` is added but unused; better deferred to the `CaseBranchMode` PR.

### Scope

The LC0095 changes (codefix rewrite ~350 lines, HasFix/HasFixAll fixtures, resx entries, instruction files) are entirely unrelated to FC0007 and should be split into their own PR — for reviewability and independent revertability.

### Verdict

The core FC0007 implementation is well-tested for the paths it covers, but shouldn''t merge as-is: fix the else-branch FP (High) and the null-settings crash, reconcile blank-line semantics with the spec, and split out the LC0095 work.

Arthurvdv:
Back from the holidays! If I can somehow can assist on the PR, please let me know 🙋

MODUSCarstenScholling:
I'll need some days for internal stull. I'll come back to this in a couple of days.

MODUSCarstenScholling:
I've decided to move these changes up because they're currently blocking our transition to ALCops and an internal migration. Plus, that will free me up to focus on other things next week.

Arthurvdv:
Nice work on this rule — the control-flow and scope-leaving blank line checks cover a lot of ground. I ran a code review and spotted two interaction edge cases worth flagging for a follow-up. Since this rule isn't enabled by default, I'm comfortable shipping this with 1.1.0 and iterating based on feedback.

**1. Duplicate FC0007 when control-flow block meets scope-leaving statement**

When a control-flow block (`if`/`while`/etc.) is immediately followed by `exit` or `Error()` without a blank line, both the "after control-flow block" check and the "before scope-leaving statement" check fire on the same token — producing two FC0007 diagnostics at the same location.

```al
if Flag then begin
    Message('x');
end;
exit; // ← two FC0007 squiggles here
```

Fix idea: guard one of the two paths to skip when the other will already report (e.g., skip the ControlFlowAfter check when the next sibling is a scope-leaving statement).

**2. One-liner gap blind spot with default `OneLinerMode=None`**

A missing blank line between a multi-line control-flow block and an adjacent one-liner control-flow statement is never flagged. The multi-line block's "after" check skips because the next sibling is a control-flow statement, and the one-liner's "before" check skips because it's excluded as a one-liner. Neither fires.

```al
if Flag then begin
    Message('x');
end;
if Other then // ← missing blank line not flagged
    Message('y');
```

Fix idea: the ControlFlowAfter skip should also check whether the next sibling would actually perform its own "before" check — i.e., don't skip when the next sibling is a one-liner that will be excluded.

@MODUSCarstenScholling — would you be able to have a look at these in a follow-up PR? I'm going ahead with merging this into the release branch for 1.1.0 since the rule isn't enabled by default. No rush, just flagging while the context is fresh.

MODUSCarstenScholling:
Of course. Here we go: https://github.com/ALCops/Analyzers/pull/457

PR #457, opened 2026-08-21
fix(FC0007): resolve control-flow spacing interactions

MODUSCarstenScholling:
## Summary

Fixes FC0007 ownership of missing blank-line diagnostics when control-flow and scope-leaving checks overlap.

- Prevents duplicate FC0007 diagnostics for a control-flow block followed by `exit` or built-in `Error(...)`.
- Reports missing separators before an excluded one-line control-flow statement.
- Preserves reporting for adjacent control-flow blocks when `ControlFlowBefore` is disabled.
- Adds exact diagnostic-count coverage so duplicate diagnostics cannot regress unnoticed.
- Updates internal FC0007 rule instructions.

### DIFF
--- a/src/ALCops.FormattingCop/Analyzers/StatementBlocksSeparatedByBlankLine.cs
+++ b/src/ALCops.FormattingCop/Analyzers/StatementBlocksSeparatedByBlankLine.cs
@@ -100,7 +100,9 @@ public sealed class StatementBlocksSeparatedByBlankLine : DiagnosticAnalyzer
                 $"before '{name}' block");
         }
 
-        if (config.ControlFlowAfter && i < siblings.Length - 1 && !IsControlFlowStatement(siblings[i + 1]))
+        if (config.ControlFlowAfter &&
+            i < siblings.Length - 1 &&
+            !WillReportControlFlowBefore(siblings[i + 1], config))
         {
             ReportIfNoBlankLineBetween(
                 ctx,
@@ -165,7 +167,10 @@ public sealed class StatementBlocksSeparatedByBlankLine : DiagnosticAnalyzer
             return;
         }
 
-        var diagnostic = GetScopeLeavingSpacingDiagnostic(statement, "before scope-leaving statement 'exit'");
+        var diagnostic = GetScopeLeavingSpacingDiagnostic(
+            statement,
+            config,
+            "before scope-leaving statement 'exit'");
 
         if (diagnostic is not null)
         {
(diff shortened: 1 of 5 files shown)


src/ALCops.FormattingCop/Analyzers/StatementBlocksSeparatedByBlankLine.cs
```diff
@@ -100,7 +100,9 @@ private static void AnalyzeControlFlowStatement(
                 $"before '{name}' block");
         }
 
-        if (config.ControlFlowAfter && i < siblings.Length - 1 && !IsControlFlowStatement(siblings[i + 1]))
+        if (config.ControlFlowAfter &&
```
Arthurvdv, line 103: **Line 103** — When `OneLinerMode` is `None` (the default), a multi-line control-flow block followed by a one-liner control-flow statement will produce no diagnostic for the missing blank line.

The "after" check on line 103 skips when the next sibling is a control-flow statement (expecting that sibling's own "before" check to handle it). But if that next sibling is a one-liner, its "before" check bails out early at line 82 (`IsOneLiner && !IncludesOneLiners`). Neither side reports.

Example — no diagnostic is raised here, but a blank line is expected after the `if...end;` block:

```al
procedure Foo()
var
    Flag: Boolean;
begin
    if Flag then begin
        DoSomething();
        DoSomethingElse();
    end;
    if Flag then
        exit; // <- no blank line above, no diagnostic
end;
```

A possible fix would be to also consider whether the next sibling is actually participating:

```csharp
if (config.ControlFlowAfter && i < siblings.Length - 1
    && (!IsControlFlowStatement(siblings[i + 1])
        || (IsOneLiner(siblings[i + 1]) && !IncludesOneLiners(config))))
```

---

*Generated with Claude Opus 4.6*

MODUSCarstenScholling: Thanks for the suggestion. It correctly addresses the default case where a multi-line control-flow block is followed by an excluded one-line control-flow statement.

I implemented the same behavior through a slightly more general condition:

```csharp
!WillReportControlFlowBefore(siblings[i + 1], config)
```

This checks whether the following statement will actually run and report its own `ControlFlowBefore` check. Besides `OneLinerMode`, it also accounts for `ControlFlowBefore = false`.

With the proposed condition, a multi-line control-flow block followed by another multi-line control-flow block would still be missed when `ControlFlowBefore` is disabled: the `after` check would skip it, but the following block would not report the missing blank line either.

The helper keeps the ownership rule explicit: a gap is delegated only when the next statement is genuinely responsible for reporting it. Regression tests cover both the default one-liner case and the `ControlFlowBefore = false` configuration.

Arthurvdv: Thank you for looking more in detail into this and posting the details here.

PR #464, opened 2026-08-25
fix(fc0007,pc0038,lc0089,lc0090) Inconsistent handling of AL built-in methods that unconditionally end execution

### LINKED ISSUE #463: [Bug]: PC0038 Terminating AL methods like FieldError not treated like Error(...) in control-flow analysis (PC0038 false positive; likely affects other rules too)
### Analyzer

PlatformCop

### Issue type

### Rule ID

PC0038

### Description

ALCops appears to treat only the built-in `Error(...)` as a hard procedure terminator. The idiomatic AL alternative `Record.FieldError(...)` — which also never returns — is not recognized as terminating. That causes false positives wherever control flow depends on “this call ends the procedure.”

Confirmed for **PC0038** (*Not all code paths return a value*). The same gap likely affects other **ALCops** rules that model scope-leaving statements.

**Expected behavior**

Any ALCops analysis that treats `Error(...)` as ending a path should also treat `FieldError(...)` the same way.

For PC0038 specifically: no diagnostic when every branch either returns a value or ends in `Error(...)` or `FieldError(...)`.

**Actual behavior**

PC0038 is reported when the only non-returning path ends with `FieldError`, even though that branch cannot continue at runtime.

Replacing `FieldError` with `Error(...)` removes the diagnostic, although both have the same control-flow effect.

The [PC0038 documentation](https://alcops.dev/docs/analyzers/platformcop/pc0038/) explicitly accepts paths that end with `Error(...)`. `FieldError` should be treated the same way.

**Suggested fix**

1. Introduce (or extend) a shared helper for **unconditional terminating statements** used by control-flow analysis.
2. Include at least: `Error(...)`, `FieldError(...)` (both overloads).
3. Audit ALCops rules that currently special-case only `Error(...)` (PC0038, complexity, FC0007 scope-leaving, etc.).

### AL code to reproduce

Minimal reproduction:

```al
codeunit 50100 "PC0038 FieldError Repro"
{
    local procedure GetDescriptionFromType(SalesLine: Record "Sales Line"): Text[100]
    begin
        case SalesLine.Type of
            SalesLine.Type::Item:
                exit(SalesLine.Description);

            SalesLine.Type::"G/L Account":
                exit(SalesLine.Description);

            else
                SalesLine.FieldError(Type); // PC0038 false positive
        end;
    end;

    local procedure GetDescriptionFromType(SalesLine: Record "Sales Line"): Text[100]
    begin
        if SalesLine.Type = SalesLine.Type::Item then
            exit(SalesLine.Description)
        else
            SalesLine.FieldError(Type); // PC0038 false positive
    end;
}
```

Observed: PC0038 on GetDescriptionFromType.

Workaround that clears the diagnostic (same control flow intent):

```al

codeunit 50100 "PC0038 FieldError Repro"
{
    local procedure GetDescriptionFromType(SalesLine: Record "Sales Line"): Text[100]
    begin
        case SalesLine.Type of
            SalesLine.Type::Item:
                exit(SalesLine.Description);
            SalesLine.Type::"G/L Account":
                exit(SalesLine.Description);
            else
                Error('%1 must not be %2.', SalesLine.FieldCaption(Type), SalesLine.Type);
        end;
    end;
}
```
Observed: PC0038 is not reported.

**Why this matters**
FieldError is the idiomatic AL way to raise a field-specific validation error (correct field caption / value in the message). Forcing developers to replace it with Error(...) only to satisfy PC0038 either:

weakens the error message, or leads to unnecessary suppressions / dummy exit('') after FieldError.

**Suggested fix**
Treat FieldError as terminating statements in the PC0038 control-flow analysis, consistent with Error(...).

### Broader concern: shared “terminator” model (ALCops scope)

This looks less like a one-off PC0038 bug and more like a **shared incompleteness in how ALCops models scope-leaving / terminating statements**.

Today, ALCops seems to special-case:

- `Error(...)`
- (and often) `exit` / `exit(...)`

…but not `Record.FieldError(...)`, even though it is the standard AL idiom for a field-specific validation error that always aborts execution.

Other **ALCops** diagnostics that may need the same terminator set:

1. **PC0038** — missing return on a path (confirmed false positive with `FieldError`)
2. **Cognitive / cyclomatic complexity** (e.g. LC0089 / LC0090 / LC0009 / LC0010) — if early-exit / guard scoring treats `Error(...)` specially but ignores `FieldError`, scores can be inconsistent for equivalent guard clauses
3. **Formatting / scope-leaving rules** (e.g. FC0007) — docs describe blank-line rules for scope-leaving statements as `exit` and built-in `Error(...)`; `FieldError` is the same kind of statement from a reader’s perspective and may be handled inconsistently

**Ask:** Please centralize a list of known **unconditional** terminating AL methods and reuse it across ALCops analyzers, rather than hardcoding only `Error(...)` per rule.

For this issue, the only additional candidate beyond `Error(...)` is Record.FieldError(...). We deliberately do **not** propose adding methods like `TestField`, because those are not unconditionally terminating.

### Why this matters

`FieldError` is the idiomatic AL way to raise a field-specific validation error. Forcing `Error(...)` only to satisfy analyzers either weakens the message or pushes dummy `exit('')` / suppressions. Recognizing `FieldError` as a terminator in ALCops benefits multiple rules and keeps diagnostics consistent with how AL actually runs.

### Related Microsoft issue (AA0136 — separate report, out of scope for ALCops)

Microsoft CodeCop **AA0136** (*Do not write code that will never be hit*) shows the inverse asymmetry: it recognizes `Error(...)` as terminating for unreachable-code detection, but **not** `FieldError`. That cannot be fixed in ALCops and would need a separate report on [microsoft/AL](https://github.com/microsoft/AL).

```al
local procedure DemoUnreachable(SalesLine: Record "Sales Line")
begin
    SalesLine.FieldError(Type);
    Message('Unreachable?'); // Not flagged by AA0136

    Error('Stop');
    Message('Unreachable?'); // Is flagged by AA0136
end;
```

MODUSCarstenScholling:
Fixes the inconsistent handling of AL built-in methods that unconditionally end execution.

- Adds `FlowTerminatingBuiltIns` as the shared semantic classifier for built-in `Error` and `FieldError`.
- Updates PC0038 to treat `Record.FieldError` and `FieldRef.FieldError` as terminating return paths.
- Updates FC0007 to handle `FieldError` consistently with `Error`.
- Updates LC0089/LC0090 guard-clause handling to resolve calls semantically, preserving complexity increments for user-defined procedures named `Error` or `FieldError`.
- Supports the valid AL syntax without parentheses for parameterless `FieldRef.FieldError`.
- Uses AL case-insensitive name comparison and removes the obsolete `ThrowError` entry.
- Adds focused regression coverage and records the LC0089/LC0090 semantic-resolution design decision.

## Validation

- `CognitiveComplexity`: 18 passed
- Affected PC0038, FC0007, and Cognitive Complexity tests: 77 passed
- CI-style multi-target build (`netstandard2.1`, `net8.0`, `net10.0`): 0 errors

Fixes #463


Arthurvdv:
Thank you for creating a PR for this. 

Instead of letting Claude write a review comment I wanted to experiment with letting it create a PR, so it created four (stacked) PR's based on the code review. Would love to hear if this works for you.

MODUSCarstenScholling:
Thanks for trying this out, Arthur. I’m sorry, but I’m going to close the PR for now. My own review didn’t confirm all of the findings and found gaps, and I also accidentally merged the PRs in the wrong order. Unfortunately, I don’t have enough time at the moment to explore this workflow further, but I appreciate the experiment and the effort you put into it.

Arthurvdv:
I’m sorry, this wasn’t the outcome I was expecting. I’ll reach out to you directly so we can align and make sure we’re on the same page for future PRs.

MODUSCarstenScholling:
Looking forward to it. But the only issue here is that directly the first PR was wrong in my opinion. Like an AI suggesting the following 10 steps where step 1 already contains errors and changes all steps after.

PR #506, opened 2026-09-02
feat: add LC0100 - Global variable could be local

someC0d3r:
Implements the proposal from [Discussion #499](https://github.com/ALCops/Analyzers/discussions/499).

## Summary

Adds LC0100, `GlobalVariableCouldBeLocal`, to LinterCop. The diagnostic identifies global variables in **normal codeunits** (see reason in the end) that are used by only one procedure or trigger and whose previous object state cannot affect any reachable read.

- Categorized as `Design` with `Info` severity and disabled by default.
- Includes the AL data type in the diagnostic message, including quoted record type names.
- Uses separate wording for immutable labels without implying reinitialization.
- Provides no code action, so moving a declaration remains an explicit developer decision.

The analysis recognizes definite initialization through assignments, `Clear`, complete branches, and supported record operations such as successful `Record.Get(...)` calls. It deliberately stays silent when retained object state may matter, including reads before initialization, multiple-procedure usage, `var` arguments, unsafe recursive or reentrant paths, temporary records, reference-like types, event publishers exposing object state, and unsupported execution models.

## Tests

- 77 focused LC0100 test cases cover the discussion scenarios and additional control-flow, record, recursion, event, reentrancy, shadowing, and type-display edge cases.
- The complete LinterCop test suite passed on .NET 10: 415 passed, with 2 expected skips.
- The .NET 8 analyzer build and repository formatting verification passed against the latest `main`.
- The locally built analyzer was tested with the Microsoft Base Application and Microsoft Subcontracting App.

Companion documentation draft PR: https://github.com/ALCops/alcops.dev/pull/165

---

**Base Application:**
Filtered on 'Label':
image

Filtered on everything else than 'Label':
image

**Subcontracting:**
Filtered on 'Label':
image

Filtered on everything else than 'Label' (have added cases myself; nothing found in the original source):
image

---

This implementation and draft PR were created with Codex using **GPT-5.6 Sol** with **Ultra** reasoning.

---

**Reason: Why only codeunits for the moment?**

LC0100 is intentionally restricted to normal codeunits in its initial implementation. Other AL object types have framework-controlled lifecycles: triggers may be invoked repeatedly or re-entered by the platform, and global variables may be consumed through declarative properties rather than ordinary procedure code. Therefore, having all procedural references inside a single trigger does not necessarily mean that procedure-local lifetime is appropriate.

For example, a report trigger is invoked once per data item record, and a global variable may intentionally retain state between those invocations:

```al
report 50100 "Sales Totals"
{
    dataset
    {
        dataitem(SalesLine; "Sales Line")
        {
            trigger OnAfterGetRecord()
            begin
                RunningTotal += SalesLine.Amount;
            end;
        }
    }

    var
        RunningTotal: Decimal;
}
```

Page globals can also be read implicitly by the UI through declarative bindings:

```al
page 50100 "Customer Example"
{
    PageType = Card;
    SourceTable = Customer;

    layout
    {
        area(Content)
        {
            field(StatusText; StatusText)
            {
                ApplicationArea = All;
            }
        }
    }

    trigger OnAfterGetCurrRecord()
    begin
        StatusText := StrSubstNo('Customer %1', Rec."No.");
    end;

    var
        StatusText: Text;
}
```

In this case, the runtime consumes `StatusText` outside the procedure operation tree. Treating pages, tables, reports, and similar objects exactly like codeunits could therefore introduce false positives. Support for additional object types can be added later once their lifecycle, declarative consumers, and reentrancy behavior are modeled and covered by dedicated tests.

### DIFF
--- a/.claude/rules/diagnostics/lc0100-global-variable-could-be-local.md
+++ b/.claude/rules/diagnostics/lc0100-global-variable-could-be-local.md
@@ -35,7 +35,7 @@ Reports global variables in normal codeunits whose references are confined to on
 - Registers a compilation action and visits each declared application object independently.
 - Collects source global variables, resolves every identifier through the semantic model, and groups eligible variables by their single containing method or trigger.
 - Binds one operation tree per method and evaluates all candidates together with a flow-sensitive `OperationWalker`.
-- Reports at the global declaration and includes the variable and method names in the diagnostic message.
+- Reports at the global declaration and includes the AL type, variable name, and method name in the diagnostic message. Labels use wording that does not imply runtime reinitialization.
 
 ## Known issues
 
--- a/.claude/rules/diagnostics/lc0100-global-variable-could-be-local.md
+++ b/.claude/rules/diagnostics/lc0100-global-variable-could-be-local.md
@@ -35,7 +35,7 @@ Reports global variables in normal codeunits whose references are confined to on
 - Registers a compilation action and visits each declared application object independently.
 - Collects source global variables, resolves every identifier through the semantic model, and groups eligible variables by their single containing method or trigger.
 - Binds one operation tree per method and evaluates all candidates together with a flow-sensitive `OperationWalker`.
-- Reports at the global declaration and includes the AL type, variable name, and method name in the diagnostic message. Labels use wording that does not imply runtime reinitialization.
+- Reports at the global declaration and includes the variable name, AL type, and method name in the diagnostic message. Labels use wording that does not imply runtime reinitialization.
 
 ## Known issues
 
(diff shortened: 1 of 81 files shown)


Arthurvdv:
@someC0d3r, I was planning to review your PR, but I see you've closed it. Is this by accident or do you want to create a new PR?

someC0d3r:
> @someC0d3r, I was planning to review your PR, but I see you've closed it. Is this by accident or do you want to create a new PR?

Ah sorry, should have mentioned it also in the PR here:
> The PRs are only a "proof-of-concept" - so nothing I would currently want & suggest to merge into main..
> (... also the PRs **got closed by me instantly**, so they don't appear on the PR list)

I didn't want to bloat up the PR list by my Draft PRs.. :D

I'll reopen it so you can have a look and leave some comments:)

someC0d3r:
According to discussion https://github.com/ALCops/Analyzers/discussions/499 this new rule will be updated. For the mean time this PR will be closed and a new one will be created once the implementation is finished.

PR #507, opened 2026-09-03
fix(DC0007,DC0008,DC0004): document interfaces and control add-ins

### LINKED ISSUE #451: [Bug]: DC0007/DC0008 never fire for interface and controladdin objects
### Analyzer

DocumentationCop

### Issue type

False negative (diagnostic doesn't fire when it should)

### Rule ID

DC0007 / DC0008

### Description

While fixing #438 (DC0004 false positive on interface procedures), a related latent bug surfaced in `ObjectRequiresDocumentation`:

The analyzer registers for the `Interface` and `ControlAddIn` symbol kinds, but the analysis callback bails out early:

```csharp
if (ctx.Symbol is not IApplicationObjectTypeSymbol appObjectTypeSymbol)
{
    return;
}
```

In the NAV SDK, `IInterfaceTypeSymbol` and `IControlAddInTypeSymbol` implement only `IObjectTypeSymbol`, **not** `IApplicationObjectTypeSymbol` (the two interfaces are siblings). So interface and controladdin objects **never** raise DC0007/DC0008, even though the registration suggests they were intended to be covered.

### Question for the original author

What is the intended object-level behavior?

1. Should public interfaces raise DC0007 (and internal ones DC0008)? Interface signatures are the cross-extension API contract, so this seems consistent with the rule intent.
2. Should controladdin objects be covered too, or removed from the registration? For DC0004 we decided to skip controladdin *procedures* entirely (JS-implemented contract), but the object itself is referenced cross-extension via `usercontrol`.

### Additional context

- Root cause analysis and the layered-resolution fix pattern are documented in `.github/instructions/dc0004-procedure-requires-documentation.instructions.md` (added by the PR fixing #438).
- Related: #438

MODUSCarstenScholling:
## Summary

Fixes #451

- Restore DC0004 diagnostics for undocumented control add-in procedures.
- Report undocumented control add-in event declarations as DC0009 with event-specific wording.
- Remove the unavailable `local`/`internal` remediation from the public-procedure diagnostic.
- Extend DC0007 and DC0008 object analysis to interfaces and control add-ins by analyzing `IObjectTypeSymbol`.
- Preserve the test-codeunit exemption by applying it only to symbols that also implement `IApplicationObjectTypeSymbol`.
- Report DC0007 for undocumented public interfaces and control add-ins, and DC0008 for undocumented internal interfaces.
- Add regression coverage for documented and undocumented interface/control-add-in objects, plus parameterized control-add-in events.
- Remove the vacuous internal-procedure control-add-in fixture.
- Document the corrected control-add-in procedure/event diagnostic routing in the analyzer and documentation repositories.


Arthurvdv:
**Review of #507 — document interfaces and control add-ins**

Solid, well-scoped fix; the object-level `IObjectTypeSymbol` widening and the separate `EventDeclaration` action are the right shapes, and the fixtures/docs are updated. A few things worth addressing before merge:

**1. DC0004 message is inaccurate for control add-in events (and procedures).** `AnalyzeControlAddInEvents` reports `PublicProcedureRequiresDocumentation`, so an undocumented `event ControlAddInReady()` surfaces as:

> Public **procedure** 'ControlAddInReady()' must include XML documentation comments (or be restricted to local or internal scope).

It is not a procedure, and control add-in members take no `local`/`internal` modifier, so the remediation is impossible. `DC0009` (`EventRequiresDocumentation`, "Event '{0}' ...") is enabled by default and reads correctly for events. Worth reconsidering the ID for events, or at least a control-add-in-appropriate message. (The same "restricted to local or internal scope" tail is now emitted for control add-in *procedures* too.)

**2. Missing NoDiagnostic regression for documented interfaces/control add-ins (object rule).** `ObjectRequiresDocumentation` gained `PublicHasDiagnostic`/`InternalHasDiagnostic` fixtures only. There is no fixture proving a *documented* public/internal interface or control add-in is left alone, so a future false positive on these `IObjectTypeSymbol`-only kinds would go uncaught.

**3. The parameterized-event path is untested.** Every control add-in event fixture is parameterless, so `GetEventDisplayText`/`GetTypeDisplayText` — including the `#if NETSTANDARD2_1` reflection branch — never execute under test. A `HasDiagnostic` fixture with `event Foo(bar: Integer)` would cover it.

**4. `InternalNoDiagnostic/ControlAddIn.al` now passes vacuously.** It still holds an undocumented `procedure MyProcedure()` and asserts no DC0006. With the exemption removed, that procedure now emits DC0004, not DC0006 — so the test stays green without validating anything meaningful. Consider updating or removing it to reflect the new behavior.

**5. Minor: display-text duplication.** `GetEventDisplayText`/`GetTypeDisplayText` re-implement what the method path already gets from `GetDiagnosticDisplayText(MethodSymbolDisplayFormat.MethodSignature)`; the two formatters can drift over time.

Nothing here is a crash; items 1–2 are the ones I'd want resolved before merge.

Arthurvdv:
I think we’ve covered everything now, so let’s merge this into main. And again thank you for creating the PR 🤗

PR #540, opened 2026-09-11
fix(DC0007): detect XML documentation on profiles

frottke:
## What happened

I found a false positive in DC0007 for documented `profile` objects. The profile had XML documentation directly above its declaration, but the analyzer still reported that documentation was missing.

The issue is specific to how the NAV SDK exposes documentation for profiles. `GetDocumentationCommentXml()` does not reliably return the comment in this case, even though the documentation is present in the source.

## What changed

The analyzer now checks the leading syntax trivia of the object declaration for XML documentation comments. This correctly recognizes both single-line and multi-line documentation comments.

The existing profile handling remains unchanged. Profiles are still analyzed, and undocumented profiles continue to raise DC0007.

## Tests

I added regression fixtures for both cases:

- a documented profile that must not raise DC0007
- an undocumented profile that must raise DC0007


Arthurvdv:
Thank you! 🙏

